PR #164, opened 2023-02-09
SDK-3949: Introduce sending client information with requests

ewanharris:
### 🔧 Changes

Introduces telemetry to the package which is being sent in the (new) `Auth0-Client` header.

By default the telemetry data sent is for the go-auth0 package however, it can be overridden by passing a custom Telemetry to the `WithTelemetry` function on `API.New`. The structure is as follows:
* name: A string that is name of the "thing" the telemetry data is for
* version: A string that is version of the "thing" the telemetry data is for
* env: A object of strings that is a grab-bag for any other data that the consumer of go-auth0 wishes to send, like OS version for example

By calling `WithNoTelemetry` the `Auth0-Client` header is not sent at all

### 🔬 Testing

I'm working on figuring out how to test this to make sure the data shows up as expected (rather than works in theory)

### 📝 Checklist

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -23,24 +23,24 @@ import (
 // UserAgent is the default user agent string.
 var UserAgent = fmt.Sprintf("Go-Auth0-SDK/%s", auth0.Version)
 
-// Telemetry is the structure used to send telemetry data in the "Auth0-Client" header.
-type Telemetry struct {
+// Auth0ClientInfo is the structure used to send client information in the "Auth0-Client" header.
+type Auth0ClientInfo struct {
 	Name    string            `json:"name"`
 	Version string            `json:"version"`
 	Env     map[string]string `json:"env,omitempty"`
 }
 
-// IsEmpty checks whether the provide Telemetry data is nil or has no data to allow
+// IsEmpty checks whether the provided Auth0ClientInfo data is nil or has no data to allow
 // short-circuiting the "Auth0-Client" header configuration.
-func (td *Telemetry) IsEmpty() bool {
+func (td *Auth0ClientInfo) IsEmpty() bool {
 	if td == nil {
 		return true
 	}
 	return td.Name == "" && td.Version == "" && len(td.Env) == 0
 }
 
-// DefaultTelemetryData is the default telemetry data sent by the go-auth0 SDK.
-var DefaultTelemetryData = &Telemetry{Name: "go-auth0", Version: auth0.Version}
+// DefaultAuth0ClientInfo is the default client information sent by the go-auth0 SDK.
+var DefaultAuth0ClientInfo = &Auth0ClientInfo{Name: "go-auth0", Version: auth0.Version}
 
 // RoundTripFunc is an adapter to allow the use of ordinary functions as HTTP
 // round trips.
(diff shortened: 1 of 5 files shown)


internal/client/client.go
```diff
@@ -14,12 +14,34 @@ import (
...
+	"encoding/json"
+
 	"github.com/auth0/go-auth0"
 )
 
 // UserAgent is the default user agent string.
 var UserAgent = fmt.Sprintf("Go-Auth0-SDK/%s", auth0.Version)
 
+// Telemetry is the structure used to send telemetry data in the "Auth0-Client" header.
+type Telemetry struct {
```
Widcket, line 27: While some of our older SDKs still call this Telemetry, more recently we've started to call it UserAgent instead, as "telemetry" can often have negative connotations and raise eyebrows.
For example: https://github.com/auth0/Auth0.Android/pull/398

Please consider renaming the feature.

Widcket: Other example:
https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter/lib/auth0_flutter.dart#L27

internal/client/client.go
```diff
@@ -14,12 +14,34 @@ import (
...
+	Name    string            `json:"name"`
+	Version string            `json:"version"`
+	Env     map[string]string `json:"env,omitempty"`
+}
+
+// IsEmpty checks whether the provide Telemetry data is nil or has no data to allow
```
Widcket, line 33:
```suggestion
// IsEmpty checks whether the provided Telemetry data is nil or has no data to allow
```

internal/client/client.go
ewanharris: I think that's a fair point, unfortunately this package already exposes a [configurable user agent](https://github.com/auth0/go-auth0/blob/main/internal/client/client.go#L21) so I'm not sure I can rename it to that, in SPA-JS it's called [`Auth0Client`](https://auth0.github.io/auth0-spa-js/interfaces/Auth0ClientOptions.html#auth0Client), what do you think about that?

Widcket: Is that user agent string public? If not, you can change that as well. Otherwise, the SPA SDK solution is fin 👍🏼

ewanharris: Configuring the  user agent is public, by calling `WithUserAgent` on a `management.New` call and is used by [auth0 cli](https://github.com/auth0/auth0-cli/blob/e4cab13cebf1b949b15c018d36357663db609a26/internal/cli/cli.go#L212) and [terraform](https://github.com/auth0/terraform-provider-auth0/blob/f11337f3818810e9fde04be184b46ff3f623b855/internal/provider/provider.go#L133) currently

internal/client/client.go
ewanharris: Resolved in c336a83a92ef478bbe2e7de232ed68a0751f5431

internal/client/client.go
ewanharris: Pushed c336a83a92ef478bbe2e7de232ed68a0751f5431 to change this, I ending up using `Auth0ClientInfo` to try and convey that this is information about the client rather than a client that can be used to call API methods (if that makes sense)

ewanharris:
CI is failing currently because of changes in golanglint-ci as we always pull the latest. Will investigate that on Monday but solutions are either pin to a specific golanglint-ci version or update to use go 1.19 (at a minimum)

Widcket (approved):

PR #175, opened 2023-03-09
Add support for Ping Federate connection

ewanharris:
### 🔧 Changes

This adds support for the Ping Federate connection, it does not expose the `expires`, `subject`, or `thumbprints` properties as these are populated on creation/update by the backend based on the provided cert (which can be a string representing the contents of a pem file, or a base64 representation of the contents of a pem file).

In v1 we could potentially look into providing read-only properties such as these in a way that disallows a developer from overriding them.

### 📚 References

terraform-provider-auth0/issues/428

### 🔬 Testing

Test creating and updating the properties of a Ping Federate connection locally alongside the unit tests added

### 📝 Checklist


Widcket (dismissed):

sergiught (changes requested):

sergiught (approved):
💪🏻

PR #190, opened 2023-04-04
Allow passing extra environment data to the auth0client data

ewanharris:
### 🔧 Changes

Adds a new `WithAuth0ClientEnvOption` func so that users of the SDK can enhance the env data with their own version to allow us to see their total usage, this is mostly for use in auth0-cli and terraform so that we can distinguish between their usage of the SDK and customer usage

### 🔬 Testing

Covered by unit tests and also covered by manual testing and verifying the API log in manage

### 📝 Checklist

### DIFF
--- a/management/management_option.go
+++ b/management/management_option.go
@@ -93,6 +93,18 @@ func WithAuth0ClientInfo(auth0ClientInfo client.Auth0ClientInfo) Option {
 	}
 }
 
+// WithAuth0ClientEnvOption allows adding extra environment keys to the client information.
+func WithAuth0ClientEnvOption(key string, value string) Option {
+	return func(m *Management) {
+		if !m.auth0ClientInfo.IsEmpty() {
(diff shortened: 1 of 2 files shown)


management/management_option.go
```diff
@@ -93,6 +93,18 @@ func WithAuth0ClientInfo(auth0ClientInfo client.Auth0ClientInfo) Option {
 	}
 }
 
+// WithAuth0ClientEnvOption allows adding extra environment keys to the client information.
+func WithAuth0ClientEnvOption(key string, value string) Option {
```
Widcket, line 97: Just curious, why call it "Option"?

ewanharris: I wanted to try and convey this was to add to the env object rather than replace it, maybe `WithAuth0ClientEnvEntry` fits better? I wasn't sure on whether `Key` or `Value` fully suited

Widcket (approved):

Issue #189, opened 2023-03-30
List log endpoints returns scope field as an array of strings instead of string for certain log types.


mscno:
### Checklist

- [X] I have looked into the [README](https://github.com/auth0/go-auth0#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [documentation](https://pkg.go.dev/github.com/auth0/go-auth0) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/go-auth0/issues) and have not found a suitable solution or answer.
- [X] I have upgraded to the [latest version](https://github.com/auth0/go-auth0/releases/latest) of this SDK and the issue still persists.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

The list logs endpoint causes an JSON unmarshal error for certain log types:

Example of a log type that causes error (this log is extracted directly from the Auth0 management REST API):

```json
{
		"date": "2023-03-30T09:02:33.667Z",
		"type": "fsa",
		"description": "Login required",
		"client_id": "redacted",
		"client_name": "Admin",
		"ip": "79.160.79.156",
		"user_agent": "Firefox 111.0.0 / Mac OS X 10.15.0",
		"details": {
			"body": {},
			"qs": {
				"audience": "https://domain.com",
				"client_id": "redacted",
				"redirect_uri": "http://localhost:3000",
				"scope": "openid profile email",
				"response_type": "code",
				"response_mode": "web_message",
				"state": "redacted==",
				"nonce": "redacted==",
				"code_challenge": "redacted",
				"code_challenge_method": "S256",
				"prompt": "none",
				"auth0Client": "redacted="
			},
			"connection": null,
			"error": {
				"message": "Login required",
				"oauthError": "login_required",
				"type": "oauth-authorization"
			},
			"riskAssessment": null
		},
		"hostname": "tenant.eu.auth0.com",
		"audience": "https://domain.com",
		"scope": [
			"openid",
			"profile",
			"email"
		],
		"auth0_client": {
			"name": "auth0-react",
			"version": "1.11.0"
		},
		"log_id": "90020230330090235500182000000000000001223372036901203194",
		"_id": "90020230330090235500182000000000000001223372036901203194",
		"isMobile": false
	}
```

[...]

PR #194, opened 2023-04-17
Handle instances where scope is returned as array in log entries

ewanharris:
### 🔧 Changes

This implements a custom `UnmarshalJSON` function for the `Log` type that will handle the differing types of `scope`. Although this is documented to be a `string` it appears in some instances (from checking, only `fsa` type events) to be an array of strings.

Unlike other instances where we have to implement this functionality, we do not need to implement a `MarshalJSON` function as we can only read the `Log` type, i.e. there is no create or update functionality exposed.

### 📚 References

#189

### 🔬 Testing

Covered by unit tests

### 📝 Checklist

### DIFF
--- a/management/log.go
+++ b/management/log.go
@@ -1,6 +1,9 @@
 package management
 
 import (
+	"encoding/json"
+	"fmt"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


sergiught (approved):

willvedd (approved):
Great work. Will fix the `auth0 logs list` and `auth0 logs tail` commands in the Auth0 CLI.

PR #196, opened 2023-04-18
Add support for client credential management

ewanharris:
### 🔧 Changes

Adds support for the management of client credentials.

### 🔬 Testing

Covered by unit tests and also tested locally against a tenant with this enabled

### 📝 Checklist

### DIFF
--- a/management/client.go
+++ b/management/client.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strconv"
+	"time"
 
 	"github.com/auth0/go-auth0"
 )
@@ -104,6 +105,8 @@ type Client struct {
 
 	OrganizationUsage           *string `json:"organization_usage,omitempty"`
 	OrganizationRequireBehavior *string `json:"organization_require_behavior,omitempty"`
+
+	ClientAuthenticationMethods *ClientAuthenticationMethods `json:"client_authentication_methods,omitempty"`
 }
 
 // ClientJWTConfiguration is used to configure JWT settings for our Client.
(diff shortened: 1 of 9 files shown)


management/client.go
```diff
@@ -179,6 +182,40 @@ type ClientRefreshToken struct {
...
+	PEM *string `json:"pem,omitempty"`
+	// Algorithm which will be used with the credential.
+	Algorithm *string `json:"alg,omitempty"`
+	// Parse expiry from x509 certificate. If true, attempts to parse the expiry date from the provided PEM.
+	ParseExpiryFromCert *bool `json:"parse_expiry_from_cert,omitempty"`
+	// The time that this credential was created.
+	CreatedAt *time.Time `json:"created_at,omitempty"`
+	// The time that this credential was last updated.
+	UpdatedAt *time.Time `json:"updated_at,omitempty"`
+	//
```
Widcket, line 205: This is missing its documentation.

management/client.go
```diff
@@ -179,6 +182,40 @@ type ClientRefreshToken struct {
...
 
+// Credential is used to configure Client Credentials.
+type Credential struct {
+	// The ID of the credential.
+	ID *string `json:"id,omitempty"`
+	// The name of the credential
+	Name *string `json:"name,omitempty"`
+	// The key identifier of the credential.
+	KeyID *string `json:"kid,omitempty"`
+	// The credential type, only
```
Widcket, line 193: This doc seems to be incomplete.

management/client.go
ewanharris: Fixed!

management/client.go
ewanharris: Fixed!

management/client.go
```diff
@@ -240,6 +277,28 @@ func (m *ClientManager) Delete(id string, opts ...RequestOption) error {
 	return m.Request("DELETE", m.URI("clients", id), nil, opts...)
 }
 
+// CreateCredential creates an client application's client credential.
+func (m *ClientManager) CreateCredential(id string, credential *Credential, opts ...RequestOption) error {
```
Widcket, line 281: Shouldn't this return the created credential? See https://github.com/auth0/node-auth0/pull/734/files#diff-8f2f9669ed871610c0349f76ff7819e843d36da69a7818c90de89c3036ca3cddR65.

management/client.go
```diff
@@ -179,6 +182,40 @@ type ClientRefreshToken struct {
...
+	Name *string `json:"name,omitempty"`
+	// The key identifier of the credential.
+	KeyID *string `json:"kid,omitempty"`
+	// The type of credential.
+	CredentialType *string `json:"credential_type,omitempty"`
+	// PEM-formatted public key or X509 certificate.
+	PEM *string `json:"pem,omitempty"`
+	// Algorithm which will be used with the credential.
+	Algorithm *string `json:"alg,omitempty"`
+	// Parse expiry from x509 certificate. If true, attempts to parse the expiry date from the provided PEM.
```
Widcket, line 199:
```suggestion
	// Parse expiry from x509 certificate. If `true`, attempts to parse the expiry date from the provided PEM.
```

management/client.go
```diff
@@ -240,6 +277,28 @@ func (m *ClientManager) Delete(id string, opts ...RequestOption) error {
 	return m.Request("DELETE", m.URI("clients", id), nil, opts...)
 }
 
+// CreateCredential creates an client application's client credential.
+func (m *ClientManager) CreateCredential(id string, credential *Credential, opts ...RequestOption) error {
+	return m.Request("POST", m.URI("clients", id, "credentials"), credential, opts...)
+}
+
+// ListCredentials lists all client credentials associated with the client application.
+func (m *ClientManager) ListCredentials(id string, opts ...RequestOption) (c []*Credential, err error) {
```
Widcket, line 286: Should we use `clientID` here and in `CreateCredential()` to make it clearer what it is (these methods are not doing clients CRUD), and for consistency with `GetCredential()` and `DeleteCredential()`.

management/client.go
Widcket: Nevermind, it receives a pointer, so it should update in-place.

management/client.go
ewanharris: Yeah I wasn't sure what naming was best here as we seemed to have a current pattern of defaulting to `id` and then using `xID` and `yID` when there are multiple. But I agree that it's best to ensure it's obvious so I'll move to `clientID` everywhere

ewanharris: Fixed in ddf5e4081a7269917695a52062c48c69103241b6

Widcket (approved):

PR #216, opened 2023-05-30
[SDK-4224] Allow configuration and disabling of retry strategy

ewanharris:
### 🔧 Changes

This improves upon the existing retry logic implemented in the package by expanding the retry
conditions, allowing configuration of the retry conditions and allowing disabling retries.

By default the Management Client will retry a request 2 times, when there is a 429 status code
or if there is an error returned (except for a select couple of errors that are not recoverable). The `X-RateLimit-Reset` is consulted to ensure that the request will work, unless the time until the request is more than the maximum wait time.

The configuration of the max retries and status codes that will be retried on by providing a
`RetryOptions` struct with the required values with the `WithRetryOptions` function, and retries
can be disabled by passing the `WithNoRetries` function

### 📝 Checklist

### ⚠️  Breaking Change

The SDK will now retry a request a maximum of 2 times by default, and will also retry on certain errors.

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -192,7 +192,7 @@ emailProviderConfig := &management.EmailProvider{
 
 Previously, when retrying requests the SDK would retry only for 429 status codes, waiting until the time indicated by the `X-RateLimit-Reset` header and would not have a maximum number of requests (i.e would retry infinitely).
 
-Now, by default the SDK will retry 429 status codes and request errors that are deemed to be recoverable, and will retry a maximum of 2 times. The SDK will instead use an exponential backoff with a minimum time of 250 milliseconds, and a maximum delay time of 10 seconds and consult the `X-Ratelimit-Reset` header to ensure that the next request will pass.
+Now, by default the SDK will retry 429 status codes and request errors that are deemed to be recoverable, and will retry a maximum of 2 times. The SDK will instead use an exponential backoff with a minimum time of 250 milliseconds, and a maximum delay time of 10 seconds and consult the `X-RateLimit-Reset` header to ensure that the next request will pass.
 
 This logic can be customised by using the `management.WithRetries` helper, and can be disabled using the `management.WithNoRetries` method.
 
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -192,7 +192,7 @@ emailProviderConfig := &management.EmailProvider{
 
 Previously, when retrying requests the SDK would retry only for 429 status codes, waiting until the time indicated by the `X-RateLimit-Reset` header and would not have a maximum number of requests (i.e would retry infinitely).
 
-Now, by default the SDK will retry 429 status codes and request errors that are deemed to be recoverable, and will retry a maximum of 2 times. Instead of waiting for the `X-RateLimit-Reset` header value, it will instead use an exponential backoff with a minimum time of 250 milliseconds.
+Now, by default the SDK will retry 429 status codes and request errors that are deemed to be recoverable, and will retry a maximum of 2 times. The SDK will instead use an exponential backoff with a minimum time of 250 milliseconds, and a maximum delay time of 10 seconds and consult the `X-Ratelimit-Reset` header to ensure that the next request will pass.
 
 This logic can be customised by using the `management.WithRetries` helper, and can be disabled using the `management.WithNoRetries` method.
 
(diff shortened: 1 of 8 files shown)


MIGRATION_GUIDE.md
```diff
@@ -127,4 +129,28 @@ api.Client.Create()
...
+    domain,
+    management.WithClientCredentials(context.Background(), id, secret),
+    management.WithRetries(RetryOptions{MaxRetries: 3, Statuses: []int{http.StatusTooManyRequests, http.StatusBadGateway}}),
+)
+
+// Disable retries
+m, err := management.New(
+    domain,
+    management.WithClientCredentials(context.Background(), id, secret),
+    management.WithNoRetries(),
```
Widcket, line 154: I understand it might be a bit late for this, but with this design it would be possible to use both `WithRetries()` and `WithNoRetries()` –even though the last used method would prevail. What about simply using `management.WithRetries(RetryOptions{MaxRetries: 0})` ?

Widcket: Consider as well using factory methods if necessary, to reduce the verbosity.

Widcket: Also, what is the SDK already doing with other similar configuration options?

ewanharris: Yeah I was a little undecided on this and initially had an `Enabled` property on the `RetryOptions` struct.

We currently have `WithNoAuth0ClientInfo` to disable the Auth0Client telemetry so that was part of the reason in going for `WithNoRetries` for this API

internal/client/client.go
```diff
@@ -58,32 +84,89 @@ func (rf RoundTripFunc) RoundTrip(req *http.Request) (*http.Response, error) {
...
+		backoffDelay(),
+	)
+
+	return tr
 }
 
-func retry(attempt rehttp.Attempt) bool {
-	if attempt.Response == nil {
-		return false
+// Matches the error returned by net/http when the configured number of
```
Widcket, line 117:
```suggestion
// Matches the error returned by net/http when the configured number of redirects
```

internal/client/client.go
```diff
@@ -58,32 +84,89 @@ func (rf RoundTripFunc) RoundTrip(req *http.Request) (*http.Response, error) {
...
+// is exhausted.
+var redirectsErrorRe = regexp.MustCompile(`stopped after \d+ redirects\z`)
+
+// Matches the error returned by net/http when the TLS certificate is not trusted.
+// When version 1.20 is the minimum supported this can be replaced by tls.CertificateVerificationError.
```
Widcket, line 122: Note that if you're making a new major version it is acceptable to have it support only Go 1.20+. Specially since Go 1.21 is expected for August 2023, a mere 3 months away, which will drop support for 1.19.

internal/client/client.go
```diff
@@ -58,32 +84,89 @@ func (rf RoundTripFunc) RoundTrip(req *http.Request) (*http.Response, error) {
...
-func delay(attempt rehttp.Attempt) time.Duration {
-	resetAt := attempt.Response.Header.Get("X-RateLimit-Reset")
-	resetAtUnix, err := strconv.ParseInt(resetAt, 10, 64)
-	if err != nil {
-		resetAtUnix = time.Now().Add(5 * time.Second).Unix()
+// backoffDelay implements a DelayFn that is an exponential backoff with jitter
+// and a minimum value.
+func backoffDelay() rehttp.DelayFn {
+	// Disable gosec lint for as we don't need secure randomness here and the error
+	// handling of an error adds needless complexity
```
Widcket, line 152:
```suggestion
	// handling of an error adds needless complexity.
```

MIGRATION_GUIDE.md
Widcket: Note that this being a new major version, it's an opportunity to go with the approach that makes the most sense for all methods involved (including `WithNoAuth0ClientInfo`). It's alright if you want to use `WithNoRetries()` anyway though.

internal/client/client.go
ewanharris: Yeah I think that's a fair shout, I'll maybe circle back to dropping 1.19 after checking one some of the users of this package have their compile target.

MIGRATION_GUIDE.md
ewanharris: I think given that we'd most likely have keep `WithNoAuth0ClientInfo` (i.e. adding an `Enabled` flag to that struct isn't as straight forward), then I'd lean towards maintaining the existing `WithX`/`WithNoX` pattern as it does fit with Go

Widcket (dismissed):

Widcket (approved):

PR #222, opened 2023-06-12
[SDK-4231] Implement initial structure of auth client

ewanharris:
### 🔧 Changes

Implements the initial base structure of the authentication client along with 2 methods in OAuth and Database in order to facilitate testing of JSON and form encoded requests.

This does not implement aspects such as ID token verification, telemetry and rate limiting onto the client which will be done in a future PR.

There is testing setup for these methods but currently it is dependent on tenant state (i.e. signed up users), randomising this data, recording management API interactions etc. proved difficult so I will file a ticket to ensure that this is dealt with as part of this effort. Current ideas are:

* Script that uses Auth0 CLI to set up required user
* Have a default user that will be stored for recordings and override this if running against a live tenant

### 🔬 Testing

Tested manually and covered by unit tests

### 📝 Checklist

### DIFF
--- a/authentication/authentication.go
+++ b/authentication/authentication.go
@@ -1,0 +1,61 @@
+package authentication
+
+import (
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/auth0/go-auth0/internal/client"
+)
+
+// Authentication is the auth client.
+type Authentication struct {
+	Database *Database
+	OAuth    *OAuth
+
+	url             *url.URL
+	basePath        string
+	debug           bool
+	http            *http.Client
+	auth0ClientInfo *client.Auth0ClientInfo
+	common          manager
+	clientID        string
+	clientSecret    string
+}
+
+type manager struct {
+	authentication *Authentication
+}
+
+// New creates an auth client.
+func New(domain string, options ...Option) (*Authentication, error) {
+	// Ignore the scheme if it was defined in the domain variable, then prefix
+	// with https as it's the only scheme supported by the Auth0 API.
+	if i := strings.Index(domain, "//"); i != -1 {
+		domain = domain[i+2:]
+	}
+	domain = "https://" + domain
+
+	u, err := url.Parse(domain)
+	if err != nil {
+		return nil, err
+	}
+
+	a := &Authentication{
+		url:             u,
+		basePath:        "",
+		debug:           false,
+		http:            http.DefaultClient,
+		auth0ClientInfo: client.DefaultAuth0ClientInfo,
+	}
+
+	for _, option := range options {
+		option(a)
+	}
+
+	a.common.authentication = a
+	a.Database = (*Database)(&a.common)
+	a.OAuth = (*OAuth)(&a.common)
+
+	return a, nil
+}
--- a/authentication/authentication_error.go
+++ b/authentication/authentication_error.go
@@ -26,7 +26,6 @@ func newError(response *http.Response) error {
 	// This can happen in case the error message structure changes.
 	// If that happens we still want to display the correct code.
 	if apiError.Status() == 0 {
-		// Always set status code as it is not returned from te
 		apiError.StatusCode = response.StatusCode
 		apiError.Err = http.StatusText(response.StatusCode)
 	}
--- a/authentication/authentication_error.go
+++ b/authentication/authentication_error.go
@@ -43,3 +43,32 @@ func (a *authenticationError) Error() string {
 func (a *authenticationError) Status() int {
 	return a.StatusCode
 }
+
+// UnmarshalJSON implements the json.Unmarshaler interface.
+//
+// It is required to handle the differences between error responses between the APIs.
+func (a *authenticationError) UnmarshalJSON(b []byte) error {
+	type authError authenticationError
+	type authErrorWrapper struct {
+		*authError
+		Code        string `json:"code"`
+		Description string `json:"description"`
+	}
+
+	alias := &authErrorWrapper{(*authError)(a), "", ""}
+
+	err := json.Unmarshal(b, alias)
+	if err != nil {
+		return err
+	}
+
+	if alias.Code != "" {
+		a.Err = alias.Code
+	}
+
+	if alias.Description != "" {
+		a.Message = alias.Description
+	}
+
+	return nil
+}
--- a/authentication/authentication_error.go
+++ b/authentication/authentication_error.go
@@ -1,0 +1,45 @@
+package authentication
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+type authenticationError struct {
+	StatusCode int    `json:"statusCode"`
+	Err        string `json:"error"`
+	Message    string `json:"error_description"`
+}
+
+func newError(response *http.Response) error {
+	apiError := &authenticationError{}
+
+	if err := json.NewDecoder(response.Body).Decode(apiError); err != nil {
+		return &authenticationError{
+			StatusCode: response.StatusCode,
+			Err:        http.StatusText(response.StatusCode),
+			Message:    fmt.Errorf("failed to decode json error response payload: %w", err).Error(),
+		}
+	}
+
+	// This can happen in case the error message structure changes.
+	// If that happens we still want to display the correct code.
+	if apiError.Status() == 0 {
+		// Always set status code as it is not returned from te
+		apiError.StatusCode = response.StatusCode
+		apiError.Err = http.StatusText(response.StatusCode)
+	}
+
+	return apiError
+}
+
+// Error formats the error into a string representation.
+func (a *authenticationError) Error() string {
+	return fmt.Sprintf("%d %s: %s", a.StatusCode, a.Err, a.Message)
+}
+
+// Status returns the status code of the error.
+func (a *authenticationError) Status() int {
+	return a.StatusCode
+}
--- a/authentication/authentication_error_test.go
+++ b/authentication/authentication_error_test.go
@@ -1,0 +1,61 @@
+package authentication
+
+import (
+	"io"
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_newError(t *testing.T) {
+	var testCases = []struct {
+		name          string
+		givenResponse http.Response
+		expectedError authenticationError
+	}{
+		{
+			name: "it fails to decode if body is not a json",
+			givenResponse: http.Response{
+				StatusCode: http.StatusForbidden,
+				Body:       io.NopCloser(strings.NewReader("Hello, I'm not a JSON.")),
+			},
+			expectedError: authenticationError{
+				StatusCode: 403,
+				Err:        "Forbidden",
+				Message:    "failed to decode json error response payload: invalid character 'H' looking for beginning of value",
+			},
+		},
+		{
+			name: "it correctly decodes the error response payload",
+			givenResponse: http.Response{
+				StatusCode: http.StatusBadRequest,
+				Body:       io.NopCloser(strings.NewReader(`{"statusCode":400,"error":"invalid_scope","error_description":"Scope must be an array or a string"}`)),
+			},
+			expectedError: authenticationError{
+				StatusCode: 400,
+				Err:        "invalid_scope",
+				Message:    "Scope must be an array or a string",
+			},
+		},
+		{
+			name: "it will still post the correct status code if the body doesn't have the correct structure",
+			givenResponse: http.Response{
+				StatusCode: http.StatusInternalServerError,
+				Body:       io.NopCloser(strings.NewReader(`{"errorMessage":"wrongStruct"}`)),
+			},
+			expectedError: authenticationError{
+				StatusCode: 500,
+				Err:        "Internal Server Error",
+				Message:    "",
+			},
+		},
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			actualError := newError(&testCase.givenResponse)
+			assert.Equal(t, &testCase.expectedError, actualError)
+		})
+	}
+}
--- a/authentication/authentication_error_test.go
+++ b/authentication/authentication_error_test.go
@@ -51,6 +51,19 @@ func Test_newError(t *testing.T) {
 				Message:    "",
 			},
 		},
+		{
+			name: "it will handle a invalid sign up response",
+			givenResponse: http.Response{
+				StatusCode: http.StatusBadRequest,
+				Body:       io.NopCloser(strings.NewReader(`{"name":"BadRequestError","code":"invalid_signup","description":"Invalid sign up","statusCode":400}`)),
+			},
+			expectedError: authenticationError{
+				StatusCode: 400,
+				Err:        "invalid_signup",
+				Message:    "Invalid sign up",
+			},
+		},
+	}
 
 	for _, testCase := range testCases {
 		t.Run(testCase.name, func(t *testing.T) {
--- a/authentication/authentication_error_test.go
+++ b/authentication/authentication_error_test.go
@@ -16,10 +16,10 @@ func Test_newError(t *testing.T) {
 		expectedError authenticationError
 	}{
 		{
-			name: "it fails to decode if body is not a json",
+			name: "it fails to decode if body is not json",
 			givenResponse: http.Response{
 				StatusCode: http.StatusForbidden,
-				Body:       io.NopCloser(strings.NewReader("Hello, I'm not a JSON.")),
+				Body:       io.NopCloser(strings.NewReader("Hello, I'm not JSON.")),
 			},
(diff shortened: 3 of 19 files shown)


authentication/authentication_error.go
```diff
@@ -0,0 +1,74 @@
...
+
+// Status returns the status code of the error.
+func (a *authenticationError) Status() int {
+	return a.StatusCode
+}
+
+// UnmarshalJSON implements the json.Unmarshaler interface.
+//
+// It is required to handle the differences between error responses between the APIs.
+func (a *authenticationError) UnmarshalJSON(b []byte) error {
```
ewanharris, line 50: The signature of the error returned from SignUp is different so we map it back in this unmarshal to return a consistent error

authentication/authentication_request.go
```diff
@@ -0,0 +1,201 @@
...
+// URI returns the absolute URL of the Management API with any path segments
+// appended to the end.
+func (a *Authentication) URI(path ...string) string {
+	baseURL := &url.URL{
+		Scheme: a.url.Scheme,
+		Host:   a.url.Host,
+		Path:   a.basePath + "/",
+	}
+
+	const escapedForwardSlash = "%2F"
```
ewanharris, line 23: Should probably remove this as this is management client specific

authentication/authentication.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// New creates an auth client.
+func New(domain string, options ...Option) (*Authentication, error) {
+	// Ignore the scheme if it was defined in the domain variable, then prefix
+	// with https as it's the only scheme supported by the Auth0 API.
+	if i := strings.Index(domain, "//"); i != -1 {
+		domain = domain[i+2:]
+	}
+	domain = "https://" + domain
```
Widcket, line 37: Could we simply expect the domain to be without scheme, instead of trying to get it into shape? Unless this is implemented this way in other Management SDKs

authentication/authentication_option.go
```diff
@@ -0,0 +1,18 @@
...
+type Option func(*Authentication)
+
+// WithClientID configures the Client ID to be provided with requests if one is not provided.
+func WithClientID(clientID string) Option {
+	return func(a *Authentication) {
+		a.clientID = clientID
+	}
+}
+
+// WithClientSecret configures the Client ID to be provided with requests if one is not provided.
```
Widcket, line 13:
```suggestion
// WithClientSecret configures the Client secret to be provided with requests if one is not provided.
```

authentication/authentication.go
```diff
@@ -0,0 +1,61 @@
...
+		debug:           false,
+		http:            http.DefaultClient,
+		auth0ClientInfo: client.DefaultAuth0ClientInfo,
+	}
+
+	for _, option := range options {
+		option(a)
+	}
+
+	a.common.authentication = a
```
Widcket, line 56: Could you please expand on the rationale for having a property with a manager that points to the same object?

authentication/authentication_error.go
```diff
@@ -0,0 +1,74 @@
...
+			StatusCode: response.StatusCode,
+			Err:        http.StatusText(response.StatusCode),
+			Message:    fmt.Errorf("failed to decode json error response payload: %w", err).Error(),
+		}
+	}
+
+	// This can happen in case the error message structure changes.
+	// If that happens we still want to display the correct code.
+	if apiError.Status() == 0 {
+		// Always set status code as it is not returned from te
```
Widcket, line 29: This comment seems to be incomplete.

authentication/authentication_error.go
```diff
@@ -0,0 +1,74 @@
...
+	if err := json.NewDecoder(response.Body).Decode(apiError); err != nil {
+		return &authenticationError{
+			StatusCode: response.StatusCode,
+			Err:        http.StatusText(response.StatusCode),
+			Message:    fmt.Errorf("failed to decode json error response payload: %w", err).Error(),
+		}
+	}
+
+	// This can happen in case the error message structure changes.
+	// If that happens we still want to display the correct code.
```
Widcket, line 27: Is this a precaution we should take?

authentication/authentication_error_test.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+func Test_newError(t *testing.T) {
+	var testCases = []struct {
+		name          string
+		givenResponse http.Response
+		expectedError authenticationError
+	}{
+		{
+			name: "it fails to decode if body is not a json",
```
Widcket, line 19:
```suggestion
			name: "it fails to decode if body is not json",
```

authentication/authentication_error_test.go
```diff
@@ -0,0 +1,74 @@
...
+			givenResponse: http.Response{
+				StatusCode: http.StatusForbidden,
+				Body:       io.NopCloser(strings.NewReader("Hello, I'm not a JSON.")),
```
Widcket, line 22:
```suggestion
				Body:       io.NopCloser(strings.NewReader("Hello, I'm not JSON.")),
```

authentication/authentication_error_test.go
```diff
@@ -0,0 +1,74 @@
...
+				Body:       io.NopCloser(strings.NewReader(`{"errorMessage":"wrongStruct"}`)),
+			},
+			expectedError: authenticationError{
+				StatusCode: 500,
+				Err:        "Internal Server Error",
+				Message:    "",
+			},
+		},
+		{
+			name: "it will handle a invalid sign up response",
```
Widcket, line 55:
```suggestion
			name: "it will handle an invalid sign up response",
```

authentication/authentication_request.go
```diff
@@ -0,0 +1,201 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+// URI returns the absolute URL of the Management API with any path segments
```
Widcket, line 14:
```suggestion
// URI returns the absolute URL of the Authentication API with any path segments
```

authentication/authentication_request.go
```diff
@@ -0,0 +1,201 @@
...
+		return fmt.Errorf("failed to read the response body: %w", err)
+	}
+
+	if len(responseBody) > 0 && string(responseBody) != "{}" {
+		if err = json.Unmarshal(responseBody, &resp); err != nil {
+			return fmt.Errorf("failed to unmarshal response payload: %w", err)
+		}
+	}
+
+	return nil
```
Widcket, line 175: Since this part of both `Request` and `FormRequest` is identical, maybe we can structure the code a bit differently to avoid the duplication?

authentication/authentication_request.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// RequestOption configures a call (typically to retrieve a resource) to Auth0 with
+// query parameters.
```
Widcket, line 179:
```suggestion
// RequestOption configures a call to Auth0 with query parameters.
```
That part is only applicable to the Management API.

authentication/authentication_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+	_ "github.com/joho/godotenv/autoload"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/auth0/go-auth0/authentication/database"
+)
+
+var (
+	domain                = os.Getenv("AUTH0_DOMAIN")
+	clientID              = os.Getenv("AUTH0_AUTH_CLIENT_ID")
```
Widcket, line 20: Could the env variable be simply `AUTH0_CLIENT_ID`? Considering we have `AUTH0_DOMAIN` above.

authentication/database.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"context"
+
+	"github.com/auth0/go-auth0/authentication/database"
+)
+
+// Database manager.
+type Database manager
+
+// SignUp given a user's credentials, and a connection, will create a new user using active authentication.
```
Widcket, line 12:
```suggestion
// SignUp given a user's credentials and a connection, will create a new user using active authentication.
```

authentication/database.go
```diff
@@ -0,0 +1,23 @@
...
+//
+// This endpoint only works for database connections.
+// See: https://auth0.com/docs/api/authentication?http#signup
+func (d *Database) SignUp(ctx context.Context, params database.SignUpRequest) (r *database.SignUpResponse, err error) {
```
Widcket, line 16: Since we're using `Login` in method names, should this be `Signup` instead?

authentication/database/database.go
```diff
@@ -0,0 +1,77 @@
...
+	// The user's full name.
+	Name string `json:"name,omitempty"`
+	// The user's nickname.
+	Nickname string `json:"nickname,omitempty"`
+	// A URI pointing to the user's picture.
+	Picture string `json:"picture,omitempty"`
+	// The user metadata to be associated with the user. If set, the field must be an object containing no more than ten properties.
+	// Property names can have a maximum of 100 characters, and property values must be strings of no more than 500 characters.
+	UserMetadata *map[string]interface{} `json:"user_metadata,omitempty"`
+	// Extra parameters to be merged into the request body. Values set here will override any existing values
```
Widcket, line 34:
```suggestion
	// Extra parameters to be merged into the request body. Values set here will override any existing values.
```

authentication/database/database.go
```diff
@@ -0,0 +1,77 @@
...
+	ExtraParameters map[string]string `json:"-"`
+}
+
+// SignUpResponse is a sign up response.
+type SignUpResponse struct {
+	// The user's email address.
+	Email string `json:"email,omitempty"`
+	// Indicates whether a user has verified their email address.
+	EmailVerified bool `json:"email_verified,omitempty"`
+	// The users ID
```
Widcket, line 44:
```suggestion
	// The user's ID
```

authentication/database.go
Widcket: Same for `SignUpRequest`/`SignUpResponse`.

authentication/oauth.go
```diff
@@ -0,0 +1,80 @@
...
+// LoginWithGrant allows logging in with an OAuth 2.0 grant. This should only be needed if a grant
+// type is not supported byt this SDK.
+func (o *OAuth) LoginWithGrant(ctx context.Context, grantType string, body url.Values, opts ...RequestOption) (t *oauth.TokenSet, err error) {
+	body.Add("grant_type", grantType)
+
+	err = o.authentication.FormRequest(ctx, "POST", o.authentication.URI("oauth", "token"), body, &t, opts...)
+	return
+}
+
+// LoginWithPassword is for logging in with information is typically received from a highly trusted
```
Widcket, line 23:
```suggestion
// LoginWithPassword is for logging in with information that is typically received from a highly trusted
```

authentication/oauth.go
```diff
@@ -0,0 +1,80 @@
...
+// public client like a SPA.
```
Widcket, line 24: `a highly trusted public client like a SPA.` is this intended?

authentication/oauth.go
```diff
@@ -0,0 +1,80 @@
...
+// For single-page applications and native/mobile apps, we recommend using web flows instead.)
```
Widcket, line 25:
```suggestion
// For single-page applications and native/mobile apps, we recommend using web flows instead.
```

authentication/oauth.go
```diff
@@ -0,0 +1,80 @@
...
+//
+// See: https://auth0.com/docs/api/authentication#resource-owner-password
+//
+// Use the `Header` RequestOption to set the `auth0-forwarded-for` header to an end-users IP if you
+// you want brute force protection to work in server-side scenarios
+// See See https://auth0.com/docs/get-started/authentication-and-authorization-flow/avoid-common-issues-with-resource-owner-password-flow-and-attack-protection
```
Widcket, line 31:
```suggestion
// See https://auth0.com/docs/get-started/authentication-and-authorization-flow/avoid-common-issues-with-resource-owner-password-flow-and-attack-protection
```

authentication/oauth/oauth.go
```diff
@@ -0,0 +1,41 @@
...
+	// ClientSecret to use for the specific request .Required when Client Secret Basic or Client
+	// Secret Post is the application authentication method.
+	ClientSecret string
+}
+
+// TokenSet defines the response of the OAuth endpoints.
+type TokenSet struct {
+	// The access token.
+	AccessToken string `json:"access_token,omitempty"`
+	// The duration in seconds. that the access token is valid.
```
Widcket, line 16:
```suggestion
	// The duration in seconds that the access token is valid for.
```

authentication/oauth/oauth.go
```diff
@@ -0,0 +1,41 @@
...
+	ClientAuthentication
+	// The user's username.
+	Username string
+	// The user's password.
+	Password string
+	// String value of the different scopes the application is asking for. Multiple scopes are separated with whitespace.
+	Scope string
+	// The unique identifier of the target API you want to access.
+	Audience string
+	// String value of the realm the user belongs. Set this if you want to add realm support at this grant.
```
Widcket, line 37:
```suggestion
	// String value of the realm the user belongs. Set this if you want to add realm support to this grant.
```

authentication/authentication.go
ewanharris: This is the same as what the management SDK already does. From a quick look all SDKs have an expectation that it should be without the scheme but not all enforce it, which can lead to errors.

Java does [have some handling](https://github.com/auth0/auth0-java/blob/master/src/main/java/com/auth0/client/auth/AuthAPI.java#L148-L151), I _think_ [PHP does](https://github.com/auth0/auth0-PHP/blob/main/src/Configuration/SdkConfiguration.php#L1013-L1023), Node/Python don't touch the domain in any way

authentication/authentication.go
ewanharris: This mirrors a change that was made in the management side as part of #214 (with some inspiration from go-github), the change had 2 main reasons:

1) Prevents the chaining issue that embedding the authentication/management structs introduces (i.e. `authAPI.Database.OAuth.Databse.SignUp()`
2) A pretty small optimisation from reusing a single authentication struct allocation rather one for each manager

authentication/authentication_test.go
ewanharris: When I was testing my client id I used for the management tests wouldn't work here, I'm not sure if that was just down to me not having that client configure correctly to do auth and management calls or not. I'll investigate further and potentially stick to one env var as suggested

authentication/database.go
ewanharris: Yeah I think that's fair, will rename

authentication/oauth.go
ewanharris: I added this based on the docs in the [API docs](https://auth0.com/docs/api/authentication?http#resource-owner-password:~:text=typically%20using%20an%20interactive%20form%20in%20the%20user%2Dagent%20(browser)) and [node-auth0](https://github.com/auth0/node-auth0/blob/beta/src/auth/oauth.ts#L322)

authentication/authentication.go
Widcket: Thanks for the explanation.

authentication/oauth.go
Widcket: Just note that SPAs, being public clients, cannot hold a secret. So I'd suggest double-checking with @adamjmcgrath whether it's correct to call them "highly-trusted".

Widcket (approved):

authentication/authentication_test.go
ewanharris: I enabled all grants for the M2M app used for management tests and it doesn't seem to work for the `TestOAuthLoginWithPassword` test (however it does work for signup), I think for now it'll be best to keep the split and see if this can be improved in future.

authentication/oauth.go
adamjmcgrath: Yep - that should read "This flow should only be used from highly-trusted applications that cannot do redirects" - no idea where I got that from in node-auth0 😅  will fix

adamjmcgrath: It's "typically received" which means the un/pw is _received_ from a public client - the grant itself is _sent_ from a confidential client

Although it does read a bit strange, let me take another look at it

adamjmcgrath: I think the API docs have changed since that node-auth0 docstring and they read a bit better - I would go with that 

```
This is the OAuth 2.0 grant that highly-trusted apps use to access an API. In this flow, the end-user is asked to fill in credentials (username/password), typically using an interactive form in the user-agent (browser). This information is sent to the backend and from there to Auth0. It is therefore imperative that the application is absolutely trusted with this information. For [single-page applications and native/mobile apps](https://auth0.com/docs/flows/concepts/auth-code-pkce), we recommend using web flows instead.
```

adamjmcgrath: "Highly Trusted public client like a SPA" in this instance means, a browser based client you trust well enough to handle usernames and passwords (like lock.js) - but the updated API docs is better so I'd go with that

PR #246, opened 2023-07-20
[SDK-4412] Support validating org_name claim in ID token

ewanharris:
### 🔧 Changes

Adds validation of the `org_name` claim in an ID token, and renames the `OrganizationID` property on `IDTokenValidationOptions` to `Organization` to reflect that it can be a org ID or org name.

If the provided `IDTokenValidationOptions.Organization` value is prefixed with `org_` then it will check for existence of the `org_id` claim and check the values match, if not then it will check for the existence of the`org_name` claim and check the values match. The provided `IDTokenValidationOptions.Organization` value will be lowercased, but the `org_name` claim will not

### 📝 Checklist

### DIFF
--- a/authentication/oauth_test.go
+++ b/authentication/oauth_test.go
@@ -187,7 +187,7 @@ func TestWithIDTokenVerification(t *testing.T) {
 		assert.ErrorContains(t, err, "org_id claim value mismatch in the ID token")
 	})
 
-	t.Run("error for an invalid organization when using org_id", func(t *testing.T) {
+	t.Run("error for an invalid organization when using org_name", func(t *testing.T) {
 		extras := map[string]interface{}{
 			"org_name": "wrong-org",
 		}


authentication/oauth_test.go
```diff
@@ -173,20 +173,34 @@ func TestRevokeRefreshToken(t *testing.T) {
...
 
 		_, err = api.OAuth.LoginWithAuthCode(context.Background(), oauth.LoginWithAuthCodeRequest{
 			Code: "my-code",
-		}, oauth.IDTokenValidationOptions{OrganizationID: "right-org"})
+		}, oauth.IDTokenValidationOptions{Organization: "org_456"})
 
 		assert.ErrorContains(t, err, "org_id claim value mismatch in the ID token")
 	})
 
+	t.Run("error for an invalid organization when using org_id", func(t *testing.T) {
```
Widcket, line 190:
```suggestion
	t.Run("error for an invalid organization when using org_name", func(t *testing.T) {
```

PR #257, opened 2023-08-01
Allow using a custom object when reading users

danmcgee-soda:
The new v1 interface made the underlying `Management` client private, no longer allowing a custom API client call by users of this library.

We used to have a custom method like this that allowed us to unmarshal additional fields on the user profile, such as `locale` or `upn`, which are present for some other connection types.

	// ReadUser is a wrapper to use our CustomUser type, but is otherwise identical to
	// calling `manager.ReadUser(subject)`.
	func (a *Auth0UserClient) ReadUser(ctx context.Context, subject string) (u *CustomUser, err error) {
		err = a.Users.Request("GET", a.Users.URI("users", subject), &u, management.Context(ctx))
		return
	}

We can no longer do this with the v1 library, as the `Request` and `URI` methods are now inaccessible.

Thus, implement a new ReadCustom method that allows passing in a destination object to unmarshal to. This pattern should be familiar to anyone that has called `json.Unmarshal`. As seen in the test, you can wrap the existing user with a custom object to add any fields your particular business logic requires.

Naming is hard- I went with `ReadCustom` but am more than open to any better naming suggestions on the method name.

Alternatives I considered but discarded - also willing to change course here if desired:
* It would be possible to add the UPN and Locale struct fields to the User object. However, we see many other fields come across from our WAAD connection (tenantid, job_title, etc.), and I am unable to enumerate them in every other connection type. I used https://auth0.com/docs/manage-users/user-accounts/user-profiles/sample-user-profiles to get an idea of what other connection types might have additional properties.
* Re-expose the underlying low-level client in some way?
* Add a `OtherProperties map[string]interface{}` to the `User` object, and implement custom unmarshalling logic to put any unknown properties there. This felt error prone.

### DIFF
--- a/management/user.go
+++ b/management/user.go
@@ -399,6 +399,15 @@ func (m *UserManager) Read(ctx context.Context, id string, opts ...RequestOption
 	return
 }
 
+// Read user details for a given user_id into the provided destination. This allows access
+// to custom or non-standard user properties on the user profile.
+//
+// See: https://auth0.com/docs/api/management/v2#!/Users/get_users_by_id
+func (m *UserManager) ReadCustom(ctx context.Context, id string, destination any, opts ...RequestOption) (err error) {
+	err = m.management.Request(ctx, "GET", m.management.URI("users", id), destination, opts...)
+	return
+}
+
 // Update user.
 //
 // The following attributes can be updated at the root level:
--- a/management/user_test.go
+++ b/management/user_test.go
@@ -44,6 +44,52 @@ func TestUserManager_Read(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, expectedUser.GetID(), actualUser.GetID())
 	assert.Equal(t, expectedUser.GetName(), actualUser.GetName())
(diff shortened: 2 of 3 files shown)


ewanharris:
Hey @danmcgee-soda the change you've noticed is due to [the removal of the manager chaining](https://github.com/auth0/go-auth0/blob/main/MIGRATION_GUIDE.md#removal-of-manager-chaining). The `URI` and `Request` methods are part of the top level `management` struct, but due to how the individual resource structs were composed prior to v1 they were made available on all resource structs.

After that change, the methods are now only accessible on the top level management struct so rather than using `m.User.Request`/`m.User.URI` they are available as `m.Request`/`m.URI`. I think that should work for your custom method which aligns with our [recommendation for this scenario](https://github.com/auth0/go-auth0/blob/main/EXAMPLES.md#providing-a-custom-user-struct)

danmcgee-soda:
I totally missed that, thank you! I'll give that a try and close this out if it works.

danmcgee-soda:
Thanks for the suggestion!

PR #298, opened 2023-10-24
Add a Sort RequestOption helper

### LINKED ISSUE #221: Sorting
### Checklist

- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

I'd like to sort Users as described here: https://auth0.com/docs/api/management/v2#!/Users/get_users

### Describe the ideal solution

Having a `func Sort(s string) RequestOption { ... }` function similar to `Query` in management/management_request.go

### Alternatives and current workarounds

_No response_

### Additional context

_No response_

ewanharris:
### 🔧 Changes

Adds a `Sort` helper function that can be used to configure the field to sort by and direction.

Closes #221

### 📝 Checklist


management/management_request.go
```diff
@@ -285,3 +285,13 @@ func Stringify(v interface{}) string {
 	}
 	return string(b)
 }
+
+// Sort configures a request to sort data by the selected field.
+// Use 1 to sort ascending and -1 to sort descending.
```
Widcket, line 290: Should we use a boolean for the public API, and translate to "1" or "-1" internally? Because it's a two-state parameter.

management/management_request.go
```diff
@@ -285,3 +285,13 @@ func Stringify(v interface{}) string {
...
+func Sort(sort string) RequestOption {
```
Widcket, line 291: Should we use a string parameter for the field, and another boolean for the sort direction? Instead of expecting the exact string that the underlying API endpoint expects.

ewanharris: I'd noodled around with this a bit, initially I had `Sort("name", "-1")` then I looked at Java/.Net and they had a style like what I went with (`Sort("name:-1")`) but I'm on the fence about it as they all kinda hide what the real meaning of 1/-1 are so on a scan through usage it's not immediately obvious what's being performed.

I did consider doing `SortAscending("name")` and `SortDescending("name")` to make it clearer but figured two might be overkill.

Widcket: Agree that "1" and "-1" don't really express much about what they mean.

>I did consider doing SortAscending("name") and SortDescending("name") to make it clearer

How are we already handling similar cases in the public API, if any?
BTW these seem quite readable to me.

ewanharris: There isn't really anything too like this unfortunately, I guess the closest might the the `Query` API allowing passing [similar strings](https://github.com/auth0/go-auth0/blob/main/management/management_request.go#L241-L244)

Widcket (approved):

PR #328, opened 2023-12-08
[SDK-4715] Update documentation to state List related functions should be paginated

ewanharris:
### 🔧 Changes

Updates the doc strings for the List (and List related) functions to:
* Remove `all` to avoid indicating they do not need to be paginated
* Link to the pagination docs on pkg.go.dev

It also removes some usage of `applyListDefaults` that were added erroneously.

### 📝 Checklist

### DIFF
--- a/management/organization.go
+++ b/management/organization.go
@@ -192,6 +192,8 @@ type OrganizationManager manager
 
 // List available organizations.
 //
+// For information on how to paginate using this function see https://pkg.go.dev/github.com/auth0/go-auth0/management#hdr-Page_Based_Pagination
+//
 // See: https://auth0.com/docs/api/management/v2/#!/Organizations/get_organizations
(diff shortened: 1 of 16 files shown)


management/organization.go
```diff
@@ -393,12 +405,12 @@ func (m *OrganizationManager) AssociateClientGrant(ctx context.Context, id strin
 	}{
 		GrantID: grantID,
 	}
-	err = m.management.Request(ctx, "POST", m.management.URI("organizations", id, "client-grants"), &body, applyListDefaults(opts))
```
Widcket, line 396: Why is `applyListDefaults` incorrect here?

ewanharris: As this is a POST to associate a client grant and organization it doesn't support the parameters that `applyListDefaults` sets (such as `per_page` and `include_totals`)

Widcket: Thanks for the clarification 👍🏼

Widcket (approved):

Issue #345, opened 2024-02-07
Management User object does not support multifactor_last_modified tag


jozephp:
### Checklist

- [X] I have looked into the [README](https://github.com/auth0/go-auth0#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [documentation](https://pkg.go.dev/github.com/auth0/go-auth0) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/go-auth0/issues) and have not found a suitable solution or answer.
- [X] I have upgraded to the [latest version](https://github.com/auth0/go-auth0/releases/latest) of this SDK and the issue still persists.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

The management.User struct https://github.com/auth0/go-auth0/blob/f9765fcec82688a897503d85dce7941d2f1c1959/management/user.go#L17  currently does not support the multifactor_last_modified  which is exportable frm auth0 management API see -https://auth0.com/docs/manage-users/user-accounts/user-profiles/user-profile-structure 

### Expectation

management.User struct https://github.com/auth0/go-auth0/blob/f9765fcec82688a897503d85dce7941d2f1c1959/management/user.go#L17 should support should support multifactor_last_modified  

### Reproduction

Export an Auth0 user using management API using management.User

### Auth0 Go SDK version

v1.4.0

ewanharris:
Thanks for the issue @jozephp.

As a general rule of thumb, we try to only implement properties that are publicly documented in the [Management API docs](https://auth0.com/docs/api/management/v2/users/get-users-by-id). Additionally given that the linked document states that the property is only provided when in a user export which this SDK does not provide any functionality for handling the processing of, only requesting.

I would recommend reviewing the [example of extending the user struct](https://github.com/auth0/go-auth0/blob/main/EXAMPLES.md#providing-a-custom-user-struct) to implement this yourselves, or implement a user struct yourself.

PR #346, opened 2024-02-07
Addmultifactor last updated

PraneethaHM:
I would like to add multifactor_last_modified field inside User struct

### DIFF
--- a/management/user.go
+++ b/management/user.go
@@ -116,7 +116,7 @@ type User struct {
 	// List of multi-factor authentication providers with which this user has enrolled.
 	Multifactor *[]string `json:"multifactor,omitempty"`
 
-	// Last date of Multifactor Modified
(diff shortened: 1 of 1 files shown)


ewanharris:
Thanks for the PR @PraneethaHM.

I'm going to close this for the same reasons as #345, as a general rule of thumb, we try to only implement properties that are publicly documented in the [Management API docs](https://auth0.com/docs/api/management/v2/users/get-users-by-id). Additionally given that the linked document states that the property is only provided when in a user export which this SDK does not provide any functionality for handling the processing of, only requesting.

I would recommend reviewing the [example of extending the user struct](https://github.com/auth0/go-auth0/blob/main/EXAMPLES.md#providing-a-custom-user-struct) to implement this yourselves, or implement a user struct yourself.

PR #371, opened 2024-03-19
fix(management): don't unmarshal empty array response from request

Zarux:
This PR will make requests that return empty arrays `[]` behave the same as requests that return empty objects `{}`

### 🔧 Changes
fix: don't attempt to unmarshal empty array response in `management.Request`

### 🔬 Testing

Make a `List` request with parameters for an object that doesn't exist and the payload is a `List` struct e.g `func (m *ClientManager) List(ctx context.Context, opts ...RequestOption) (c *ClientList, err error) ` . Before this fix it could return `[]` and the unmarshal would err out, with this fix it will handle it the same as an empty object and it will not attempt to populate the payload. The payload will be unchanged

### 📝 Checklist

### DIFF
--- a/management/management_request.go
+++ b/management/management_request.go
@@ -107,7 +107,7 @@ func (m *Management) Request(ctx context.Context, method, uri string, payload in
 		return fmt.Errorf("failed to read the response body: %w", err)
 	}
 
-	if len(responseBody) > 0 && string(responseBody) != "{}" {
+	if len(responseBody) > 0 && string(responseBody) != "{}" && string(responseBody) != "[]" {
 		if err = json.Unmarshal(responseBody, &payload); err != nil {
 			return fmt.Errorf("failed to unmarshal response payload: %w", err)
 		}


ewanharris:
👋🏻  Thanks for the PR @Zarux, I'm having trouble reproducing the issue at hand. To provide some context, the expectation of this SDK is that any request sent from a `List` method should always set `include_totals` to `true` using `applyListDefaults` so we should always get the struct with totals information and not an empty array.

My concern also with this change is the return values becomes `nil` rather than a `<type>List` struct with a 0 length slice which I fear might catch folks out. Maybe this is something that could be solved with a custom unmarshaller to handle the empty array case and still return a valid struct?

Zarux:
Hi. Thanks for the reply. I am unable to reproduce as well now :thinking: This happened when doing a `List` call with defaults applied, and the `responseBody` in `management.Request` was  `[]` . But now i get the correct response so it seems this is not an issue anymore :+1:

PR #388, opened 2024-04-16
Changing UserMetadata and AppMetadata on User struct to use json.Rawmessage

JHarperFox:
### 🔧 Changes
Considering the endpoints to create and update an user in Auth0 require `app_metadata` and `user_metadata` to be JSON objects I propose switching those variables from `*map[string]interface{}` to `json.RawMessage`. 

This will allow ease of use within applications for marshalling and unmarshaling the values into structs without the overhead for dealing with type assertions or need for 3rd party libraries with deeply nested values.

### 📚 References

Create user API docs: https://auth0.com/docs/api/management/v2/users/post-users
Update user API docs: https://auth0.com/docs/api/management/v2/users/patch-users-by-id

### 📝 Checklist

### DIFF
--- a/auth0.go
+++ b/auth0.go
@@ -1,6 +1,7 @@
 package auth0
 
 import (
+	"encoding/json"
 	"fmt"
 	"time"
 )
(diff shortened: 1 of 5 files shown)


developerkunal:
Hi @JHarperFox ,

After reviewing the pull request, we've noticed that it contains breaking changes. At this time, we've decided not to proceed with integrating these changes into the current version. However, we appreciate your effort and will certainly consider incorporating them into version 2. Thank you for your work!

PR #470, opened 2024-12-05
Provider an example on correlationID

festo:
There was no example of how to add the correlation id to each request.

### 🔧 Changes
- Extend the `EXAMPLES.md` on how to add a correlation ID to the request
### 📚 References

- https://auth0.com/docs/api/management/v2#request-correlation

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -36,13 +36,46 @@ management.From()
 management.Sort()
 ```
 
+## Request Correlation
+A Correlation ID is a unique identifier (up to 64 characters) of a single Management API operation and allows for tracking such operations in tenant logs. You can find more details [here](https://auth0.com/docs/api/management/v2#request-correlation).
+
+```go
+type AddHeaderTransport struct {
+	T             http.RoundTripper
+	CorrelationID string
+}
+
+func (adt *AddHeaderTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	req = req.Clone(req.Context())
+	req.Header.Add("X-Correlation-ID", adt.CorrelationID)
+	return adt.T.RoundTrip(req)
+}
+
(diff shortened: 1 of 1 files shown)


developerkunal:
Hi @festo ,

I hope you are doing well!

Thanks for sharing your implementation! It’s a great approach for adding headers to requests. I just wanted to let you know that the SDK already supports this via `requestOpts`. For example, you can use:  

```go
requestOpts := management.Header("X-Correlation-ID", "client1_xyz")

client := &management.Client{
	Name:        auth0.String("Test Client"),
	Description: auth0.String("This is just a test client."),
	ClientMetadata: &map[string]interface{}{
		"key1": "value1",
		"key2": "value2",
	},
}

err = api.Client.Create(context.Background(), client, requestOpts)
if err != nil {
	log.Fatalf("failed to create: %+v", err)
}
log.Printf("Client: %+v", client)
```

This will include the correlation ID in the request, and it’ll show in logs under `"references": { "correlation_id": "client1_xyz" }`.  

Let us know if you have any feedback or ideas to improve this further. Appreciate your effort! 😊

festo:
Hi @developerkunal,

Thanks for sharing this, I was unaware of that solution, and missed this detail. I think it would be nice if one of the options would be documented because this feature is a bit hidden.

developerkunal:
Hi @festo,

Thank you for your input! We actually have an example of using headers here: [Request Options Example](https://github.com/auth0/go-auth0/blob/main/EXAMPLES.md#request-options).

Including all header tags in the example might make it less concise and harder to follow, so we’ve aimed to keep it focused and simple. Let us know if you have any other suggestions or feedback!

festo:
That is understandable, I should have noticed that! :) Thanks!

PR #513, opened 2025-02-20
refactor: use a single client assertion audience

panva:
To align with FAPI 2.0 and future client assertion requirements this changes the JWT Client Authentication assertion to use a single audience string value. The value itself is the issuer identifier.


developerkunal (approved):
LGTM

PR #556, opened 2025-05-16
Add `ReadEnabledClients` and `UpdateEnabledClients` methods; deprecate `EnabledClients` field in `Connection` struct of `ConnectionManager`

ErwinSteffens:
### 🔧 Changes

* Added `ReadEnabledClients` and `UpdateEnabledClients` methods to manage clients enabled on a connection.
* Deprecated the `EnabledClients` field in the `Connection` struct in favor of these explicit methods.

Refer to:

* [Update enabled clients on a connection (PATCH)](https://auth0.com/docs/api/management/v2/connections/patch-clients)
* [Read enabled clients on a connection (GET)](https://auth0.com/docs/api/management/v2/connections/get-connection-clients)

### 📚 References

* Auth0 Management API documentation on connection clients

### 🔬 Testing

* Created a client
* Created a connection
* Assigned the client to the connection using the new Go methods
* Verified the assignment in the Auth0 UI

### 📝 Checklist

* [x] All new/changed/fixed functionality is covered by tests (or N/A)
* [x] Documentation for new/changed functionality has been added (or N/A)

### DIFF
--- a/management/connection.go
+++ b/management/connection.go
@@ -253,6 +253,15 @@ func (st *SCIMToken) MarshalJSON() ([]byte, error) {
 	})
 }
 
+// UpdateEnabledClients represents the payload for updating the clients for a connection.
+type UpdateEnabledClients struct {
+	// ClientID is The client_id of the client to be the subject to change status
+	ClientID string `json:"client_id"`
+
+	// Status indicates if the connection is enabled or not for this client_id
+	Status bool `json:"status"`
+}
+
 // MarshalJSON implements the json.Marshaler interface.
 func (c *Connection) MarshalJSON() ([]byte, error) {
 	type connection Connection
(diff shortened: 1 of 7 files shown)


ErwinSteffens:
> We’d like to introduce a shared struct like this:
> 
> ```go
> type ConnectionEnabledClient struct {
>     ClientID string `json:"client_id"`
>     Status   bool   `json:"status"`
> }
> ```
> 
> This way, we can support both the `GET` and `PATCH` endpoints using a common structure.

Ok, but the `GET` endpoint just returns a list with enabled clients:

`EnabledClients *[]string `json:"enabled_clients,omitempty"``

You want me to change this? It will be a breaking change. 

Or do you just want me to rename the struct so it could be used for GET endpoint somewhere in the future?

developerkunal:
Hi @ErwinSteffens,

Thank you for your contribution!

I’ve pushed a few changes—please take a look and let me know if they look good to you.

duedares-rvj (approved):

ramya18101 (approved):

ErwinSteffens:
LGTM!

I was not aware of the endpoint to just read the enabled clients.

PR #578, opened 2025-07-25
fix: skip nil RequestOptions to prevent nil pointer dereferences on `option.apply(r)`

esotuvaka:
### 🔧 Changes

Prevents API misuse. Perhaps I should've checked the function types closer before hitting this runtime error, but this should eliminate the panic.
- When iterating `[]RequestOption`, skips when `option == nil` to prevent a runtime nil pointer dereference via `option.apply(r)`

### 🔬 Testing

Existing test suite can be extended to include passing `nil` into variadic `RequestOption` parameters on different management SDK functions, like `{SDK}.Organization.Invitations(ctx, organizationID, nil)`

### 📝 Checklist

- [x] All new/changed/fixed functionality is covered by tests 
- [x] I have added documentation for all new/changed functionality (N/A)

### DIFF
--- a/authentication/authentication_request.go
+++ b/authentication/authentication_request.go
@@ -52,7 +52,6 @@ func (a *Authentication) NewRequest(
 		}
 
 		option.apply(request, nil)
-
 	}
 
 	return request, nil
@@ -80,7 +79,6 @@ func (a *Authentication) NewFormRequest(
 		}
 
 		option.apply(request, payload)
-
 	}
 
 	body := strings.NewReader(payload.Encode())
--- a/management/actions.go
+++ b/management/actions.go
@@ -234,7 +234,6 @@ func applyActionsListDefaults(options []RequestOption) RequestOption {
 			}
 
 			option.apply(r)
-
 		}
 	})
 }
--- a/management/job.go
+++ b/management/job.go
@@ -207,7 +207,6 @@ func (m *JobManager) ImportUsers(ctx context.Context, j *Job, opts ...RequestOpt
 		}
 
 		option.apply(request)
-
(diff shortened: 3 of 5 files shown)


developerkunal:
Thanks for the explanation, it makes sense, and I agree that this is a good safety check. Skipping `nil` `RequestOptions` is a practical improvement, especially for dynamic use cases like RBAC or conditional logic.

The change looks good overall! Could you add a test to cover the `nil` case so we can make sure it stays safe going forward?

Really appreciate the contribution, thanks again!

esotuvaka:
Funny thing, the test cases were practically already there. I learned something interesting about the language today. I guess implicitly Golang interprets `options...` as an array, and with the `TestNewRequest()` function it was passing params as:

```go
request, err := api.NewRequest(
	context.Background(),
	testCase.method,
	testCase.endpoint,
	testCase.payload,
	testCase.options...)
```

Which means:
```go
api.NewRequest(
	context.Background(),
	testCase.method,
	testCase.endpoint,
	testCase.payload,
	nil...)
!=
api.NewRequest(
	context.Background(),
	testCase.method,
	testCase.endpoint,
	testCase.payload,
	nil)
```

Using values, passing `nil...` gets translated into empty slice `[]`, which caused the misleading test. This case was hidden the whole time. I created a separate iteration over test cases, with variadic and nil suffixes. Hopefully this meets project testing styles. Happy to hear any feedback.

ramya18101 (dismissed):

developerkunal:
Hi @esotuvaka ,

I overlooked the signed commits. Can you sign the commits?

PR #751, opened 2026-04-09
fix: Preserve request body in debug output for all methods

bkiran6398:
### 🔧 Changes

Fixes `DebugTransport` to preserve request bodies in debug log output for POST, PATCH, and PUT requests.

Previously, `DebugTransport` called `dumpRequest` after `base.RoundTrip()` to capture headers set by inner transports. However, `RoundTrip` consumes the request body stream, so `httputil.DumpRequestOut` found an empty body — making debug logs useless for any request with a payload.

- Buffers the request body before `RoundTrip` and restores it afterward for dumping
- GET requests and error paths continue to work correctly
- Fixes `govet` lint failure by replacing deprecated `reflect.Ptr` with `reflect.Pointer` in `internal/tag/tag.go`

### 🔬 Testing

Added 5 new unit tests to `TestDebugTransport`:

- **POST body preserved** — mock transport consumes body, asserts body appears in log
- **PATCH body preserved** — same pattern for PATCH method
- **PUT body preserved** — same pattern for PUT method
- **Body preserved on error** — verifies body is logged even when transport returns an error
- **GET with no body** — confirms nil-body path is unaffected

Run tests:
```bash
go test ./internal/client/ -run TestDebugTransport -v
```

### 📝 Checklist


developerkunal (commented):
A couple of small suggestions for the body-buffering block:

1. The original `req.Body` is never explicitly closed after `ReadAll`. It would be good to call `req.Body.Close()` before replacing it.

2. The error from `io.ReadAll` is silently discarded. If reading fails partway through, a truncated body would be sent to the server without the caller knowing. In practice this is unlikely since bodies are typically `bytes.Reader` or `strings.Reader`, but it's still worth handling.

Something like:

```go
var bodyBytes []byte
if req.Body != nil {
    bodyBytes, err := io.ReadAll(req.Body)
    req.Body.Close()
    if err != nil {
        return nil, fmt.Errorf("debug transport: failed to read request body: %w", err)
    }
    req.Body = io.NopCloser(bytes.NewReader(bodyBytes))
}
```

Otherwise the fix and tests look good!

developerkunal (approved):
LGTM

PR #816, opened 2026-07-07
fix: handle legacy string format for `customHeaders` in `ConnectionOptionsOAuth2`

### LINKED ISSUE #815: `ConnectionOptionsOAuth2` unmarshal fails on tenants with legacy string-encoded `customHeaders`
### Checklist

- [x] I have looked into the [README](https://github.com/auth0/go-auth0#readme) and have not found a suitable solution or answer.
- [x] I have looked into the [documentation](https://pkg.go.dev/github.com/auth0/go-auth0) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/go-auth0/issues) and have not found a suitable solution or answer.
- [x] I have upgraded to the [latest version](https://github.com/auth0/go-auth0/releases/latest) of this SDK and the issue still persists.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

`ConnectionOptionsOAuth2.UnmarshalJSON` fails with a type assertion error when
the Auth0 Management API returns the `customHeaders` field as a JSON-encoded
string rather than a map object.

The error message is:

```
json: cannot unmarshal string into Go struct field of type map[string]string
```

This occurs on older Auth0 tenants where the `customHeaders` value was persisted
in a legacy string format. The SDK makes no attempt to handle this variant and
instead surfaces the raw JSON decode error to the caller, which causes the
`terraform-provider-auth0` to crash when reading or refreshing an OAuth2
connection resource on such tenants.

### Expectation

When the API returns `"customHeaders": "<json-encoded-string>"`, the SDK should
not return an error. The legacy format cannot be reliably decoded without knowing
the original structure, so the field should be surfaced as `nil`/`empty` and the
operator can restore the values through their Terraform configuration.

When the API returns `"customHeaders": {"key": "value"}` (the current format),
the field should continue to be decoded into `*map[string]string` as normal.

### Reproduction

```go
package main

import (
    "encoding/json"
    "fmt"
    "github.com/auth0/go-auth0/management"
)

func main() {
    // Simulates what the Management API returns on an older tenant.
    // customHeaders is a JSON-encoded string instead of a map object.
    payload := `{
        "authorizationURL": "https://example.com/oauth/authorize",
        "tokenURL": "https://example.com/oauth/token",
        "customHeaders": "{\"X-Custom-Header\":\"custom-value\"}"
    }`

    var opts management.ConnectionOptionsOAuth2
    err := json.Unmarshal([]byte(payload), &opts)
    if err != nil {
        fmt.Printf("Error: %v\n", err)
        // Prints: json: cannot unmarshal string into Go struct field of type map[string]string
    }
}
```

Steps to reproduce in production:
 
1. Have an Auth0 tenant where an OAuth2 connection was created and had `customHeaders` set while the SDK persisted them in a legacy string format.
2. Use `terraform-provider-auth0` to run terraform refresh or terraform plan against that connection.
3. The provider crashes with the above unmarshal error.

### Auth0 Go SDK version

v1.44.0

alexmsenger:
### 🔧 Changes

Extends `ConnectionOptionsOAuth2.UnmarshalJSON` to gracefully handle the `customHeaders` field when the Auth0 Management API returns it as a JSON-encoded string rather than a `map[string]string` object.

Prior to this change, the decoder attempted to unmarshal the raw value directly into `*map[string]string`, which produced:

```
json: cannot unmarshal string into Go struct field of type map[string]string
```

The fix intercepts `customHeaders` as `interface{}` (alongside the existing `scope` handling) and normalizes the two possible forms:

- `map[string]interface{}` — decoded normally into `*map[string]string`
- `string` (legacy format) — discarded gracefully; `CustomHeaders` is left nil so the operator can restore the values through their config

**Types/methods changed:**
- `ConnectionOptionsOAuth2.UnmarshalJSON` in `management/connection.go`

### 📚 References

Closes #815

### 🔬 Testing

The existing `make test` suite passes with HTTP recordings unchanged —
the fix touches only unmarshal logic, not any API interaction.

Three new cases are added to `TestOAuth2Connection_UnmarshalJSON` in
`management/connection_test.go`, covering:

1. Normal map form — `{"customHeaders":{"X-Foo":"bar"}}` → decoded into `*map[string]string`
2. Legacy string form — `{"customHeaders":"{\"X-Foo\":\"bar\"}"}` → no error, `CustomHeaders` is nil
3. Null form — `{"customHeaders":null}` → no error, `CustomHeaders` is nil

No new HTTP recordings are required as the bug is fully exercisable without
a live tenant.

### 📝 Checklist

### DIFF
--- a/management/connection.go
+++ b/management/connection.go
@@ -1420,16 +1420,24 @@ type ConnectionOptionsOAuth2 struct {
 
 // UnmarshalJSON implements the json.Unmarshaler interface for ConnectionOptionsOAuth2.
 // It is required to handle differences in the scope field, which can
-// be an array of strings or a single string.
+// be an array of strings or a single string, and to handle the customHeaders
+// field, which legacy Auth0 tenants may return as a JSON-encoded string instead
+// of a map[string]string object.
 func (c *ConnectionOptionsOAuth2) UnmarshalJSON(data []byte) error {
 	type connectionOptionsOAuth2 ConnectionOptionsOAuth2
 
+	// RawCustomHeaders shadows the embedded CustomHeaders *map[string]string field
+	// so that encoding/json places the raw value here (as interface{}) rather than
+	// attempting to decode it directly into *map[string]string. This prevents a
+	// "cannot unmarshal string into Go struct field ... of type map[string]string"
+	// error for connections whose customHeaders were stored in a legacy format.
 	type connectionOptionsOAuth2Wrapper struct {
 		*connectionOptionsOAuth2
-		RawScope interface{} `json:"scope,omitempty"`
+		RawScope         interface{} `json:"scope,omitempty"`
+		RawCustomHeaders interface{} `json:"customHeaders,omitempty"`
 	}
 
-	alias := &connectionOptionsOAuth2Wrapper{(*connectionOptionsOAuth2)(c), nil}
+	alias := &connectionOptionsOAuth2Wrapper{(*connectionOptionsOAuth2)(c), nil, nil}
 
 	err := json.Unmarshal(data, alias)
 	if err != nil {
@@ -1449,6 +1457,29 @@ func (c *ConnectionOptionsOAuth2) UnmarshalJSON(data []byte) error {
 			c.Scope = auth0.String(rawScope)
 		default:
 			return fmt.Errorf("unexpected type for field scope: %T", alias.RawScope)
+		}
+	}
+
+	// Normalise customHeaders: accept both the current map[string]interface{} form
+	// and the legacy string form (a JSON-encoded value from an older schema version).
+	// In the legacy string case we discard the value — it cannot be reliably decoded
+	// without knowing the original structure — so custom_headers will appear empty
+	// and can be set manually in the Terraform config.
+	if alias.RawCustomHeaders != nil {
+		switch v := alias.RawCustomHeaders.(type) {
+		case map[string]interface{}:
+			headers := make(map[string]string, len(v))
+			for key, val := range v {
+				if s, ok := val.(string); ok {
+					headers[key] = s
+				}
+			}
+			if len(headers) > 0 {
+				c.CustomHeaders = &headers
+			}
+		case string:
+			// Legacy format: discard. The provider will surface custom_headers as
+			// empty; the operator can restore the values in their Terraform config.
 		}
 	}
 
--- a/management/connection_test.go
+++ b/management/connection_test.go
@@ -1630,12 +1630,32 @@ func TestOAuth2Connection_UnmarshalJSON(t *testing.T) {
 		`{"scope":null}`:                {Scope: nil},
 		`{}`:                            {},
 		`{"scope":[]}`:                  {Scope: auth0.String("")},
+		`{"customHeaders":{"X-Foo":"bar","X-Baz":"qux"}}`: {
+			CustomHeaders: &map[string]string{"X-Foo": "bar", "X-Baz": "qux"},
+		},
 	} {
 		var actual *ConnectionOptionsOAuth2
 		err := json.Unmarshal([]byte(expectedAsString), &actual)
 		assert.NoError(t, err)
 		assert.Equal(t, expected, actual)
 	}
+
+	t.Run("It handles a legacy string-encoded customHeaders value without error", func(t *testing.T) {
+		// Some older Auth0 tenants stored customHeaders as a JSON-encoded string
+		// rather than a map. The unmarshaler must not return an error in this case;
+		// instead it discards the value so the caller can restore it via config.
+		var actual *ConnectionOptionsOAuth2
+		err := json.Unmarshal([]byte(`{"customHeaders":"{\"X-Foo\":\"bar\"}"}`), &actual)
+		assert.NoError(t, err)
+		assert.Nil(t, actual.CustomHeaders)
+	})
+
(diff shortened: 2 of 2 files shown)


management/connection.go
```diff
@@ -1469,6 +1477,29 @@ func (c *ConnectionOptionsOAuth2) UnmarshalJSON(data []byte) error {
 		}
 	}
 
+	// Normalise customHeaders: accept both the current map[string]interface{} form
+	// and the legacy string form (a JSON-encoded value from an older schema version).
+	// In the legacy string case we discard the value — it cannot be reliably decoded
+	// without knowing the original structure — so custom_headers will appear empty
+	// and can be set manually in the Terraform config.
+	if alias.RawCustomHeaders != nil {
```
rforen, line 1485: wsl: "ranges should only be cuddled with assignments used in the iteration". `headers := make(...)` on the line above isn't the variable being ranged over (`v` is), so wsl requires a blank line here.

```suggestion

			for key, val := range v {
```

management/connection.go
```diff
@@ -1469,6 +1477,29 @@ func (c *ConnectionOptionsOAuth2) UnmarshalJSON(data []byte) error {
...
+		switch v := alias.RawCustomHeaders.(type) {
+		case map[string]interface{}:
+			headers := make(map[string]string, len(v))
+			for key, val := range v {
+				if s, ok := val.(string); ok {
```
rforen, line 1491: wsl: "if statements should only be cuddled with assignments". This `if` is cuddled with the closing `}` of the `for` loop above, not an assignment, so wsl requires a blank line here.

```suggestion

			if len(headers) > 0 {
```

management/connection.go
```diff
@@ -1469,6 +1477,29 @@ func (c *ConnectionOptionsOAuth2) UnmarshalJSON(data []byte) error {
...
+					headers[key] = s
+				}
+			}
+			if len(headers) > 0 {
+				c.CustomHeaders = &headers
```
rforen, line 1497: wsl: "block should not end with a whitespace (or comment)". This `case string:` branch has no statement, only a comment, and this repo's `.golangci.yml` doesn't set `allow-trailing-comment` (default `false`), so wsl always flags a case ending in nothing but a comment. Adding a no-op `_ = v` gives the branch an actual (behavior-preserving) statement and clears the lint error without changing behavior.

```suggestion
			// empty; the operator can restore the values in their Terraform config.
			_ = v
```

rforen (commented):
The fix is functionally correct, but CI's `wsl` (whitespace linter, golangci-lint v2.5.0) job is failing with 3 violations in the new `UnmarshalJSON` code. I don't have push access to this fork branch, so here are ready-to-apply suggestions — I verified each one locally by running `golangci-lint run -c .golangci.yml ./...` (the exact command this repo's `make lint` uses) against the patched file, and confirmed it now reports **0 issues** for the whole repo.

Summary of the 3 violations and fixes:
1. `for key, val := range v {` (originally line 1489) — wsl's `range cuddled without use` rule: a `for`/`range` can only be cuddled with an assignment directly above it that is *used in the range expression itself* (`v`, not `headers`). Fix: blank line before the `for`.
2. `if len(headers) > 0 {` (originally line 1494) — same cuddle rule for `if`: it's cuddled with the `for` loop's closing brace, not an assignment. Fix: blank line before the `if`.
3. `}` closing the `switch` (originally line 1500) — wsl's `block should not end with a whitespace (or comment)` rule. The `case string:` branch's only content is a comment (no statement), and wsl's default `AllowTrailingComment` is `false` in this repo's `.golangci.yml` (no override is configured), so a case that ends in nothing but a comment is always flagged. Fix: add a no-op `_ = v` statement after the comment so the case has an actual (behavior-preserving) statement.

All three fixes are whitespace/no-op only — no logic changes.

rforen:
Disclaimer: the review above (suggested wsl lint fixes) was generated with the help of Claude (Anthropic's AI assistant), on my behalf and reviewed before posting.

rforen:
@alexmsenger would you be able to review PR https://github.com/alexmsenger/go-auth0/pull/1  to pass all checks in this PR

rforen:
@developerkunal , fyi the following test has been run locally: 
- Ran `terraform plan -target='module.auth0_spa_app["OUR_APP"].module.social_connections["OUR_CONNECTION"]'` with the stock v1.57.0 provider — confirmed it crashes with the exact bug: cannot unmarshal string into Go struct field ConnectionList.connections.connectionOptionsOAuth2.customHeaders of type map[string]string.
- Re-ran the identical targeted plan with the patched provider (built from alexmsenger:fix/oauth2-legacy-custom-headers-unmarshal, PR #816) — succeeded cleanly, `Plan: 1 to import, 86 to add`, no unmarshal error.

Is there anything else I can do to help progress this PR?

alexmsenger:
> @alexmsenger would you be able to review PR [alexmsenger#1](https://github.com/alexmsenger/go-auth0/pull/1) to pass all checks in this PR

Hi @rforen 👋 

Thank you for the review, I think the code is up to date now, correct? Anything else I can do to support?

rforen (approved):

PR #825, opened 2026-07-14
fix: apply default query params when a paginated List receives a nil request

JohnRoesler:
### 🔧 Changes

Calling a paginated `List` method with a `nil` request pointer silently dropped its default query parameters. For endpoints that default `include_totals: true`, `include_totals` was never sent; the API then defaulted it to `false` and returned a bare JSON array instead of the totals-wrapped object, so decoding failed with:

```
json: cannot unmarshal array into Go value of type management.unmarshaler
```

**Root cause:** `internal.QueryValuesWithDefaults` returned early before applying defaults whenever there was no struct to inspect (nil interface or typed-nil pointer), so a `nil` request produced an empty query string. This change applies the provided defaults directly in that case, so `nil` requests still receive the intended `page`, `per_page`, and `include_totals` values.

Because the fix lives in a shared, generator-independent helper, it applies once and covers all callers:

- **68** generated `List` clients call `QueryValuesWithDefaults`; all now apply defaults for `nil` requests.
- **27** of those default `include_totals: true` and previously hit the crash above with a `nil` request — all fixed by this change (e.g. `users/authenticationmethods`, `clients`, `roles`, `resourceservers`, `logs`, `usergrants`, `devicecredentials`, `rules`, `hooks`, `networkacls`, `keys/encryption`, `forms`, `flows/*`, `selfserviceprofiles`, `prompts/rendering`, and the `organizations/*` / `users/*` families).
- The remaining **41** never crashed (they don't select response shape via `include_totals`) but now correctly honor their `page`/`per_page` defaults on `nil` requests too.

**Known limitation (out of scope):** explicitly passing `IncludeTotals: false` to one of the 27 shape-dependent endpoints still returns a bare array the offset pager cannot decode. An offset pager needs `total` to paginate, so honoring the explicit value is correct. Fully closing this would require the generated paginated response types to accept both the object and array shapes in `UnmarshalJSON`, a larger generator-level change left as a possible follow-up.

Files changed:

- `management/internal/query.go` — apply defaults when no request struct is present.
- `management/internal/query_test.go` — update the nil-input case (`nil input applies defaults`) to assert defaults are applied for both `nil` and typed-nil-pointer inputs.
- `management/users/authenticationmethods/.../users_authentication_methods_regression_test.go` — new hand-written (non-generated) regression test.

### 📚 References

Fixes #<issue-number>

### 🔬 Testing

- Added a hand-written regression test using `httptest` that calls `List(ctx, "id", nil)`, asserts `include_totals=true` is sent, and verifies the object response decodes. The mock returns a bare array when `include_totals != true`, so the test fails loudly if the default is ever dropped again.
- Verified the new test fails with the reported error (`json: cannot unmarshal array into Go value of type management.unmarshaler`) when the fix is reverted, and passes with it.
- `go test ./management/internal/...` passes.
- Full suite via `make test` (Docker WireMock) and `make lint`.

### 📝 Checklist


JohnRoesler:
@developerkunal could you or someone on your team take a look?

JohnRoesler:
@ramya18101 @duedares-rvj @developerkunal

developerkunal:
Hey @JohnRoesler,

I hope you're doing well!

I'll raise this issue with Fern, as this should be generated automatically rather than updated manually. Thanks for taking the time to contribute and point this out!

developerkunal:
Hi @JohnRoesler, following up on this. Before anything else, `test_package/main.go` in this PR contains what look like real credentials (a client secret and a subject JWT) along with a live staging host. Could you please remove that file? I'd also strongly recommend rotating that client secret on your side, because it will stay reachable in the branch history even after the file is deleted.

As I mentioned earlier, this fix belongs in our Fern generation pipeline rather than a hand edit, so we're going to reproduce the change there and open a separate PR with the equivalent fix. Because of that, I'll go ahead and close this one, but the fix itself will ship through the generated PR.

Thanks again for surfacing this and for the clear reproduction. It made the issue easy to act on.

PR #840, opened 2026-07-30
feat: Add support for `Token Vault Privileged Access` 

KartikJha:
# feat(client): add Token Vault privileged access support

### 🔧 Changes

Adds support for configuring a client as a **Token Vault privileged worker**, which allows it to request Token Vault tokens on behalf of other users. This is an Early Access feature and must be enabled for the tenant.

**Types added**

- `ClientTokenVaultPrivilegedAccess` — the configuration object itself, with three fields:
  - `Credentials *[]ClientCredentialID` — credentials attached to the worker (max 2). Required whenever the object is being set, on both `POST /clients` and `PATCH /clients/{id}`. Sending an empty array detaches every credential and disables the worker.
  - `IPAllowlist *[]string` — permitted IPv4/IPv6 addresses or CIDR ranges the worker may request tokens from (max 10).
  - `Grants *[]ClientTokenVaultPrivilegedGrant` — pins the connections, and the scopes within them, the worker may request tokens for (max 5 grants, max 20 scopes in total).
- `ClientCredentialID` — references a client credential. On `PATCH /clients/{id}` only `ID` is accepted, referencing a credential created beforehand via `ClientManager.CreateCredential`; on `POST /clients` the credential material is supplied instead through `CredentialType` and `PEM`.
- `ClientTokenVaultPrivilegedGrant` — a `Connection` name plus the `Scopes` allowed on it.

**Fields added**

- `Client.TokenVaultPrivilegedAccess *ClientTokenVaultPrivilegedAccess` (`token_vault_privileged_access`).

**Why the sub-fields are pointers to slices**

Each sub-field is a `*[]T` so all three write semantics of the Management API can be expressed distinctly:

| Value | Sent as | Effect |
| --- | --- | --- |
| `nil` | key omitted | stored value left unchanged |
| `&[]T{}` | `[]` | stored value cleared |
| `&[]T{...}` | `[...]` | stored value replaced |

Removing the whole object is a fourth case that requires a literal `null` for `token_vault_privileged_access`, which `omitempty` on a nil `Client` field cannot emit. That case is documented on the field and needs a custom `PATCH` request:

```go
type clientPatch struct {
	TokenVaultPrivilegedAccess *ClientTokenVaultPrivilegedAccess `json:"token_vault_privileged_access"` // no omitempty
}

err := api.Request(ctx, http.MethodPatch, api.URI("clients", clientID), &clientPatch{})
```

**`CleanForPatch` behaviour**

`Client.CleanForPatch` now reduces a read-back credential list to bare IDs, since `PATCH /clients/{id}` only accepts credential references. Entries without an ID are deliberately left untouched rather than dropped — dropping them would silently detach the worker's credentials, whereas passing them through surfaces the API error. Credential material supplied for a create (`CredentialType`/`PEM`) is therefore preserved as-is, and an empty credentials list stays empty.

**Usage**

```go
credential, err := api.Client.CreateCredential(ctx, clientID, &Credential{
	Name:           auth0.String("Worker Key"),
	CredentialType: auth0.String("public_key"),
	PEM:            auth0.String(publicKeyPEM),
})

err = api.Client.Update(ctx, clientID, &management.Client{
	TokenVaultPrivilegedAccess: &management.ClientTokenVaultPrivilegedAccess{
		Credentials: &[]management.ClientCredentialID{{ID: credential.ID}},
		IPAllowlist: &[]string{"10.0.0.1", "192.168.1.0/24"},
		Grants: &[]management.ClientTokenVaultPrivilegedGrant{
			{
				Connection: auth0.String("google-oauth2"),
				Scopes:     &[]string{"https://www.googleapis.com/auth/calendar.readonly"},
			},
		},
	},
})
```

**Generated code**

Regenerated getters and generated tests in `management.gen.go` / `management.gen_test.go` for the new types and field.

### 📚 References

Auth0 Management API: `token_vault_privileged_access` on `POST /api/v2/clients` and `PATCH /api/v2/clients/{id}` (Early Access).

### 🔬 Testing

- `TestClientTokenVaultPrivilegedAccess_MarshalJSON` — unit tests over the (un)marshalling contract: nil sub-fields omitted, empty arrays emitted, populated arrays emitted, the create-time credential material shape, and unmarshalling a stored object.
- `TestClientTokenVaultPrivilegedAccess_CleanForPatch` — unit tests that read-back credentials are reduced to IDs, credential material is preserved, and an empty credentials list survives.
- `TestClient_TokenVaultPrivilegedAccess` — HTTP-recorded end-to-end test (`test/data/recordings/TestClient_TokenVaultPrivilegedAccess.yaml`) that creates a non-interactive client, creates a credential, attaches the object via `PATCH`, then asserts each write semantic in turn: an unrelated field update does not clobber the stored object, a populated array replaces while a nil sub-field is omitted, an empty array clears, and a custom `PATCH` with a literal `null` removes the object entirely.

Run against a tenant with the feature enabled:

```bash
make test-integration FILTER="TestClient_TokenVaultPrivilegedAccess"
```

Reviewers can replay the recordings without a tenant:

```bash
go test ./management/... -run 'TestClient.*TokenVaultPrivilegedAccess'
```

Not covered: server-side validation of the documented limits (2 credentials, 10 allowlist entries, 5 grants / 20 scopes) and the non-empty / no-duplicate scope rules — these are enforced by the API, not the SDK. The `POST /clients` path that inlines credential material is exercised only at the marshalling level, since the integration test must create credentials after the client exists in order to reference them.

### 📝 Checklist

### DIFF
--- a/management/client.go
+++ b/management/client.go
@@ -282,18 +282,21 @@ type Client struct {
 
 // ClientTokenVaultPrivilegedAccess configures Token Vault privileged worker access for a client.
 //
-// All three fields are required whenever this object is being written, on both
-// POST /clients and PATCH /clients/{id}. Leaving any one of them nil omits its
-// key and is rejected with a 400, so a caller must always send the full intended
-// state rather than only the sub-fields that changed:
+// All three fields are required on every write, on both POST /clients and
+// PATCH /clients/{id}. Omitting any one of them is rejected with a 400, so a
+// caller must always send the full intended state rather than only the
+// sub-fields that changed.
 //
-//	400 Updating `token_vault_privileged_access` requires `grants` to be present.
+// Writing this object also requires the create:client_token_vault_privileged_access
+// or update:client_token_vault_privileged_access scope, alongside the matching
+// client_credentials scope.
 //
-// Each field is a pointer to a slice so that clearing can be distinguished from
-// omission:
+// Each field is a pointer to a slice so that an empty array can be
+// distinguished from an omitted key:
 //
 //   - &[]T{...} → replaces the stored value.
-//   - &[]T{}    → sends an empty array, clearing that sub-field.
+//   - &[]T{}    → sends an empty array. PATCH clears the value, while POST
+//     rejects an empty IPAllowlist or Grants.
 //   - nil       → omits the key, which is an error (see above).
 //
 // Removing the whole object requires sending a literal `null` for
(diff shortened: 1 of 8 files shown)


management/client.go
```diff
@@ -263,6 +263,90 @@ type Client struct {
...
+
+// ClientTokenVaultPrivilegedGrant pins the scopes of a single connection that a
+// Token Vault privileged worker is allowed to request tokens for.
+type ClientTokenVaultPrivilegedGrant struct {
+	// Connection is the name of the connection the grant applies to. The
+	// connection does not need to exist when the grant is configured; it is
+	// validated at runtime.
+	Connection *string `json:"connection,omitempty"`
+
+	// Scopes the privileged worker may request on the connection.
```
developerkunal, line 347: Shouldn't these two be non-pointer types since they're required fields?

KartikJha: Need to check if a pointer variable with null values honours the `omitempty` tag

developerkunal (approved):
LGTM

PR #847, opened 2026-08-10
feat: Add support for managing Org level roles

duedares-rvj:
Adds organization-level role support to the `/api/v2/roles` endpoints (EA only)

### 🔧 Changes

| Field | JSON | Notes |
| --- | --- | --- |
| `Type *string` | `type,omitempty` | `tenant` (default) or `organization` |
| `OwnerID *string` | `owner_id,omitempty` | Owning organization ID; null for tenant-level roles |

Both fields flow through `Create`, `Read`, `Update`, and `List` with no manager-method changes. The new `type`/`owner_id` list filters work via the existing `Parameter(...)` option, and checkpoint pagination (`from`/`take`/`next`) is already covered by `Take`/`From` and the embedded `List` struct.

The two new organization endpoints in the spec (`.../roles/{role_id}/members` and `.../groups`) are intentionally out of scope here and will land in v2.
### 🔬 Testing

- `TestRoleManager_OrganizationLevelRole` covers all  changed endpoints. Recordings added.

### 📝 Checklist

### DIFF
--- a/management/management.gen.go
+++ b/management/management.gen.go
@@ -12337,6 +12337,22 @@ func (r *Role) GetName() string {
 	return *r.Name
 }
 
+// GetOwnerID returns the OwnerID field if it's non-nil, zero value otherwise.
+func (r *Role) GetOwnerID() string {
+	if r == nil || r.OwnerID == nil {
+		return ""
+	}
+	return *r.OwnerID
+}
+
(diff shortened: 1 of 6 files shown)


developerkunal (approved):
LGTM

PR #863, opened 2026-09-04
feat: add B2B integration configuration to client

bkiran6398:
### 🔧 Changes

Adds `B2BIntegrationConfiguration` support to `Client`, letting a client be configured as a B2B Integration (Enterprise Connect) client.

- New `B2BIntegrationConfiguration` type with `IntegrationType` and `SSOProfiles` fields.

### 🔬 Testing

Covered by generated accessor unit tests.

### 📝 Checklist

### DIFF
--- a/management/client.go
+++ b/management/client.go
@@ -196,6 +196,11 @@ type Client struct {
 	// https://github.com/auth0/go-auth0/blob/main/EXAMPLES.md#providing-a-custom-user-struct
 	OrganizationDiscoveryMethods *[]string `json:"organization_discovery_methods,omitempty"`
 
+	// B2BIntegrationConfiguration configures the client as a B2B Integration client
+	// (Enterprise Connect). It is set at create time: it cannot be added to, or removed
+	// from, an existing client via PATCH, only its contents updated.
+	B2BIntegrationConfiguration *B2BIntegrationConfiguration `json:"b2b_integration_configuration,omitempty"`
+
 	// AsyncApprovalNotificationChannels is an array of notification channels for contacting
 	// the user when their approval is required.
 	//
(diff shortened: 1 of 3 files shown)


management/client.go
```diff
@@ -196,6 +196,11 @@ type Client struct {
 	// https://github.com/auth0/go-auth0/blob/main/EXAMPLES.md#providing-a-custom-user-struct
 	OrganizationDiscoveryMethods *[]string `json:"organization_discovery_methods,omitempty"`
 
+	// B2BIntegrationConfiguration configures the client as a B2B Integration client
+	// (Enterprise Connect). It is set at create time: it cannot be added to, or removed
+	// from, an existing client via PATCH, only its contents updated.
+	B2BIntegrationConfiguration *B2BIntegrationConfiguration `json:"b2b_integration_configuration,omitempty"`
```
KartikJha, line 202:
```go
// It is set at create time: it cannot be added to, or removed
// from, an existing client via PATCH, only its contents updated.
```

The spec says sending `null` **does** clear (remove) the object. 

The restriction is only that it cannot be *added* via PATCH to a client that was created without it. 

Is the "cannot be removed" claim in comment correct ?

management/client.go
```diff
@@ -196,6 +196,11 @@ type Client struct {
...
+
 	// AsyncApprovalNotificationChannels is an array of notification channels for contacting
 	// the user when their approval is required.
```
KartikJha, line 205: The spec explicitly says to remove `"b2b_integration"` as a valid `app_type` value (introduced in beta) and tag it as deprecated. But the `AppType` field comment in `client.go` still lists all values without any deprecation note. Is it intentional ?

KartikJha (changes requested):

management/client.go
bkiran6398: `null` value cannot be passed in PATCH:
**Request**
```http
PATCH /api/v2/clients/2WyOlpyjwrrf7WgrZQOEQkMqTQQUu6Iw
Authorization: Bearer <redacted>
Content-Type: application/json

{"b2b_integration_configuration":null}
```

**Response — HTTP 400**
```json
{
  "statusCode": 400,
  "error": "Bad Request",
  "message": "Payload validation error: 'Expected type object but found type null' on property b2b_integration_configuration (Configuration for B2B Integration clients).",
  "errorCode": "invalid_body"
}

management/client.go
bkiran6398: `b2b_integration` was never added to SDK  as a valid `app_type`, hence no need to deprecate it.

KartikJha: Cool

management/client.go
KartikJha: Cool

KartikJha (approved):
LGTM

